PR #136, opened 2020-10-24
GitHub Action fixes

fpetru:


StephanStS:
@MichaIng, @fpetru: Shall we keep the .wordlist in alphabetical order, or do we want new words appended at the end?  
I would prefer alphabetical, then I can find similar words at one place.

fpetru:
@StephanStS - yes, alphabetical order is perfect. I will make a commit to update my latest additions.

PR #333, opened 2021-02-05
Enhanced description of dietpi-survey

StephanStS:


docs/dietpi_tools.md
```diff
@@ -426,11 +426,63 @@ It is one of the core tools, enabling you to install or uninstall one or more [*
...
+    - No private data is sent. No one can identify you. No IP address is obtained.
+    - This data allows the DietPi project team to focus and improve areas based on popularity. Ensuring the most common devices and software titles will receive the most support and improvements.
+    - By opting out, you are potentially preventing DietPi from achieving the best future possible experience for everyone.
+
+    #### How often does DietPi Survey data get sent?
+
+    The DietPi Survey data file is sent when you install software with `dietpi-software` and update DietPi.
+
+    #### How big is the data upload?
+    The data we upload is tiny (1 KB) and won't effect your internet bandwidth or system performance.
```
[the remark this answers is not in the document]
MichaIng: Correct

docs/dietpi_tools.md
```diff
@@ -426,11 +426,63 @@ It is one of the core tools, enabling you to install or uninstall one or more [*
...
-    This option will upload information on a anonymous base. The aggregated results of all uploads can be seen on the [DietPi survey statistics website](https://dietpi.com/survey/).  
-    Run `dietpi-survey`.
+    #### What is DietPi Survey?
+
+    DietPi Survey allows the DietPi project to obtain general information regarding your system and installed software.
 
     ![DietPi-Survey screenshot](assets/images/dietpi-survey.jpg)
 
+    - No private data is sent. No one can identify you. No IP address is obtained.
+    - This data allows the DietPi project team to focus and improve areas based on popularity. Ensuring the most common devices and software titles will receive the most support and improvements.
```
MichaIng, line 436: I'm thinking if this is the right way to say it. It may sound like new or small projects won't have a chance and that we're doing like the big mobile app stores, shown/pushing what is already mainstream and dropping the alternative projects. Of course it is not unreasonable and the other way round it makes much less sense, but I personally do not have a look into the survey statistics to decide whether I should work on a software title or not, or how much time I'm gonna invest, but more based on issues/bug reports, thought-though suggestions of users and simply when I'm working around those code bits already and see that something is perfectly fine, or outdated, or does not yet follow the standards we're implementing now.

For me it is more for general interest and to make the really hard decisions: when it's about dropping support for software that is either not maintained anymore, or does not support modern frameworks/runtime systems, or support from Debian side has decreased. E.g. a current topic is software with Python 2 support, in the past it was software which did not yet support PHP7.3, while others dropped support for PHP7.0. So when it starts to be required to do nasty workarounds, custom builds or fixes and such things. That time with PHP7.0 -> PHP7.3 migration on Stretch was due to the high ownCloud usage numbers, while for the low amount of phpBB users (and there was one or two other software titles) I wouldn't have done it. And I won't start to hack Python 2 into Debian Bullseye for software that doesn't want or cannot migrate to Python 3, if there is not a notable large interest in it.

Thanks to the survey I know that we can remove the Jessie support branch and migration code, as there are no Jessie systems active anymore, which is good. I also see that dropping Stretch support this summer might become a problem, so I'm currently rethinking this plan. Similarly I now know that we need to faster offer Bullseye images for ALL SBCs, so that we do not create more Buster systems when it is already oldstable, so that the current issue with Stretch will hopefully be better two years later with Buster.

And I know that the Firefly image is actually not reasonable to keep, since there are zero reported users now. However, it is very easy to update, and I see downloads from time to time. And in such case I would love to have more than 14% survey participants to be more certain about the actual number, since it's statistically totally possible that actually there are 20 or more Firefly users. It also could be zero, and for sure it's nothing compared to other SBCs, but if 5 interactive minutes (I use unattended runtime for other things), hopefully soon a completely automated build, serve accumulated 20 people with an up-to-date server image, it's fine for me 😉.

Long text, sorry, I just want to give some insights about how I personally use the survey, but I'm not sure how to put this best into one sentence, or if other contributors use it differently 😉.

docs/dietpi_tools.md
```diff
@@ -426,11 +426,63 @@ It is one of the core tools, enabling you to install or uninstall one or more [*
...
+    - By opting out, you are potentially preventing DietPi from achieving the best future possible experience for everyone.
```
MichaIng, line 437: I think it would be good to to use "opt out" here, since in fact the survey is an "opt-in" system, so by default no data is sent, only when the user interactively decides to do so. Also in the script, we use both, have an opt-in and opt-out button, but actually that may cause confusion 😅.

Another point. By not opting in, especially the own hardware and software usage is not recognised, so back to the hard decisions above, opting in helps to assure that we keep up support for software and hardware that you use.

docs/dietpi_tools.md
```diff
@@ -426,11 +426,63 @@ It is one of the core tools, enabling you to install or uninstall one or more [*
...
+
+    The DietPi Survey data file is sent when you install software with `dietpi-software` and update DietPi.
+
+    #### How big is the data upload?
+    The data we upload is tiny (1 KB) and won't effect your internet bandwidth or system performance.
+
+    #### What Data is sent?
+    In the command `dietpi-survey` the exact copy of the data we receive can be viewed. The file we receive is written in bash code to allow us faster report page creation for <https://dietpi.com/survey>.
+
+    Following is an example how this file may look like. It is from a very basic system, so only two software packages (#103, #104) are installed.
```
MichaIng, line 449: It's Dropbear and DietPi-RAMlog btw. Shall we add the info that `dietpi-software list | grep ' =2'` can be used to get the IDs, names and info of software installed via `dietpi-software`?

docs/dietpi_tools.md
```diff
@@ -426,11 +426,63 @@ It is one of the core tools, enabling you to install or uninstall one or more [*
...
+    By staying Opted in, you are supporting the DietPi project with no impact to your system or private data.
+    You can opt out of DietPi Survey by running the following command:
+
+    ```sh
+    dietpi-survey
+    ```
+
+    #### How do I check my current Opted status?
+
+    Simply run the program. The current mode will be pre-selected and highlighted.
```
MichaIng, line 480: > or use `grep 'SURVEY_OPTED_IN' /boot/dietpi.txt` to derive the status directly from your `dietpi.txt`.

docs/dietpi_tools.md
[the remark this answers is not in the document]
StephanStS: @ravenclaw900: Thanks for the feedback, good hint. I will change it generally from "How bis is the data upload?" to "Uploaded data size" resp. "The data we upload is ..." to "Uploaded data is tiny (about 1 KB)..." for all the points.

@MichaIng: I will restructure the whole "DietPi tools" page later. For now only the contents is important.

docs/dietpi_tools.md
StephanStS: Ok, added this info.

docs/dietpi_tools.md
StephanStS: @MichaIng:  
Changed it to positive description: 

*Basically DietPi survey is an ***opt OUT*** system. By selecting ***opt IN***, you are helping us achieving the best future possible experience for everyone and hat we keep up support for software and hardware that you use.*

Please review.

docs/dietpi_tools.md
StephanStS: OK, added this description.

docs/dietpi_tools.md
StephanStS: OK, changed the text.

@MichaIng: Please also review the sentence describing "Data transmission events": Is it `dietpi-software` and `dietpi-update` or also `apt upgrade` resp. `apt dist-upgrade` resp. `apt full-upgrade` resp. `apt-get ...` etc?

MichaIng: It's at the end of `dietpi-software` installs and `dietpi-update`, so all correct 👍.

docs/dietpi_tools.md
```diff
@@ -424,12 +424,100 @@ It is one of the core tools, enabling you to install or uninstall one or more [*
...
+        - ensure the most common devices and software titles will receive support and improvements,
+        - keep up support for software and hardware that you use.
+
+    #### Data transmission events
+
+    The DietPi Survey data file is sent when you install software with `dietpi-software` and update DietPi with `dietpi-update`.
+
+    #### Uploaded amount of data
+
+    The uploaded data is tiny (about 1 KB) and won't effect your internet bandwidth or system performance.
```
ravenclaw900, line 450: It still says effect, instead of affect.

docs/dietpi_tools.md
MichaIng: Ah, äh actually it's the wrong way round now 😄. DietPi is an "Opt-in" system, which means that you need to opt in to have data sent.

So also below:
```
By staying ***opt IN***, you are supporting...
```
=>
```
By selecting ***opt IN***, you are supporting ... by running the following command and follow the instructions:
```
and
```
You can ***opt OUT*** of DietPi Survey by running the following command and follow the instructions:
```
=>
```
You can always revert by selecting ***opt OUT*** to stop send data, which will as well clear the data file that you may have uploaded already.
```
or something like this.

docs/dietpi_tools.md
StephanStS: Oh, thanks. I misunderstood your remark. Of course, affect is right.

docs/dietpi_tools.md
StephanStS: @MichaIng: Ok, next try.

I moved text to an admonition at the top of the DietPi Survey chapter:

`-----<snip>-----`
grafik
`-----<snip>-----`

I try to keep track that I also change the related comment within `dietpi.txt` from phpBB to the docs page.

MichaIng (approved):

PR #628, opened 2021-12-02
Enable new tabs style with scroll buttons

MichaIng:


MichaIng:
Looks and works fine now IMO:

tabs

The right button is currently hovered by mouse pointer (hidden in screenshot).

There are some other minor but nice visual enhancements, e.g. to info/warn/notes boxes. The large "Previous"/"Next" navigation at the bottom has gone. It can be re-enabled, if wanted, but given that the same can be reached via left-side navigation, it is probably not needed? https://squidfunk.github.io/mkdocs-material/setup/setting-up-the-footer/#navigation

I also added a "to top" button.

PR #633, opened 2021-12-05
Functional extensions of dietpi-backup described

StephanStS:


fpetru:
@MichaIng - Let's include also this one in the new merge to `main`

MichaIng:
Note when changing the image dimensions, those need to be changed in every case in Markdown where this image is used.

MichaIng (approved):

PR #653, opened 2021-12-31
Fix and align some x86_64 docs parts

MichaIng:


docs/install.md
```diff
@@ -173,7 +173,7 @@ Select the following tabs for the installation description of your target.
 
     ![VirtualBox appliance import screenshot](assets/images/dietpi-VirtualBox-import1.png){: width="483" height="308" loading="lazy"}
 
-    In the following dialog the user has to choose DietPi_VirtualBox-x86_64-Bullseye.ova as the file which shall be imported.
+    In the following dialog the user has to choose DietPi_VirtualBox-x86_64-Buster.ova as the file which shall be imported.
```
MichaIng, line 176: Here it needs to stay `Buster` until the image is updated. Or we change it to:
```
...has to choose the `.ova` file which shall be imported.
```
?

StephanStS (approved):

docs/install.md
```diff
@@ -244,11 +244,11 @@ Select the following tabs for the installation description of your target.
             VBoxManage guestproperty set "<vm_name>" "/VirtualBox/GuestAdd/VBoxService/--timesync-set-threshold" 10000
             ```
 
-            Replace `<vm_name>` with the name of your virtual machine (i.e. the name shown in the VM VirtualBox Manager UI, e.g. `DietPi_VirtualBox-x86_64-Bullseye` if you import the VM without changing its name).
+            Replace `<vm_name>` with the name of your virtual machine, shown in the VM VirtualBox Manager UI, e.g. `DietPi_VirtualBox-x86_64-Bullseye` if you import the VM without changing its name.
 
         With all these setup steps the time synchronization with the usage of the extension pack should work. Sometimes it needs a couple of minutes until the time is synchronized, so be somehow patient.
 
-=== "VMware Player/Fusion"
+=== "VMware"
```
MichaIng, line 251: There are so many different VMware products, let's not try to list them all. Test and fix for ESXi is still outstanding, but generally it shall work on all of them.

docs/install.md
```diff
@@ -695,8 +695,8 @@ Select the following tabs for the installation description of your target.
 
         <font size="+2">1. Download and extract the DietPi direct write image</font>
 
-        Download the **BIOS direct write image** `DietPi_NativePC-BIOS-x86_64-Bullseye.7z` from [`dietpi.com`](https://dietpi.com/#download) and
-        unzip the downloaded file to a local folder. It is a _7z_ archive format so you will need to install either [7zip for Windows](https://www.7-zip.org/) or other alternative tools.
+        Download the **Native PC for BIOS/CSM** > **Direct flash Image** from [`dietpi.com`](https://dietpi.com/#download) and
```
MichaIng, line 698: On the download page it is called `Direct flash Image`, in the docs it is called `Direct write image`. Shall we align this? Actually `write` fits better as it is more generic, while `flashing` fits more for writing to an internal ROM or so, though common for SD cards for some reason.

StephanStS:
Like you proposed, write seems better than flash.

-----Original Message-----

MichaIng:
I opened a PR to change the wording on the download page as well: https://github.com/MichaIng/DietPi-Website/pull/185

PR #700, opened 2022-02-28
Add Proxmox install docs

MichaIng:


docs/install.md
```diff
@@ -314,6 +313,76 @@ Select the following tabs for the installation description of your target.
...
+        apt update
+        apt full-upgrade
+        apt install p7zip
+        curl -O https://dietpi.com/downloads/images/DietPi_Proxmox-x86_64-Bullseye.7z
+        7zr x DietPi_Proxmox-x86_64-Bullseye.7z
+        # Verifying the SHA256 hash
+        sha256sum -c < <(mawk '/SHA256/{print $2"  DietPi_Proxmox-x86_64-Bullseye.qcow2"}' hash.txt)
+        # Import the disk image
+        ID=100 # replace "100" with your VM ID
+        qm importdisk "$ID" DietPi_Proxmox-x86_64-Bullseye.qcow2 local-lvm
```
StephanStS, line 372: The ID given here should match to the ID given in a screenshot above (see remark to line 344).

docs/install.md
```diff
@@ -314,6 +313,76 @@ Select the following tabs for the installation description of your target.
...
+        qm set "$ID" --boot order=scsi0
+        ```
+
+    - You can now remove the downloaded and extracted files and close the console:
+
+        ```sh
+        rm DietPi_Proxmox-x86_64-Bullseye.* hash.txt README.md
+        ```
+
+    The VM can now be started, select it via left side navigation of the Proxmox web interface, then the **Start** button at the top right side, finally the **Console** button to watch and finish the DietPi first run setup.
```
StephanStS, line 383: An additional screenshot would be good: ["A picture is worth a thousand words."](https://en.wikipedia.org/wiki/A_picture_is_worth_a_thousand_words)

StephanStS (changes requested):

docs/install.md
```diff
@@ -314,6 +313,76 @@ Select the following tabs for the installation description of your target.
...
+
+    <font size="+2">1. Generate a new Proxmox VM</font>
+
+    - Access the Proxmox web interface via HTTPS on TCP port **8006**:
+        - URL: `https:<your.IP>:8006`
+        - Username: `root`
+        - Password: `<root user password your entered during Proxmox VE install>`
+    - Select the **Create VM** button at the top right corner. Choose a **Node** , a **VM ID** and a **Name**, then click **Next**.
+    - Select "**Do not use any media** ", for the **Guest OS** assure that **Linux** and version "**5.x - 2.6 Kernel**" is selected, then click **Next**.
+    - As **Machine** you can select `q35`, but the older default will work as well. We recommend the para-virtualised **VirtIO SCSI** controller, which should be the default. Click **Next**.
```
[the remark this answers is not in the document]
MichaIng: I thought about this, but the step would need to include a lot of screenshots as of the individual tabs through the VM creation. Not sure if this is really helpful, given the vast need for scrolling and that the textual instructions need to be self-explaining in any case.

docs/install.md
```diff
@@ -314,6 +313,76 @@ Select the following tabs for the installation description of your target.
...
+    <font size="+2">Introduction</font>
+
+    Virtual machine images are great for those occasions where you want to set up a DietPi system very quickly and test things. Also it may be used as a Debian based Linux system with a small footprint for development purposes, e.g. with the X11 window system. The small footprint makes it optimally usable on PCs without a huge built in RAM. Also several VMs may be run for different applications.
+
+    One big advantage of such a VM is that it needs only a couple of minutes coming to a running DietPi system.
+
+    One of the options to run and manage virtual machines is [__Proxmox__](https://www.proxmox.com/).
+
+    ![Proxmox logo](assets/images/proxmox-logo.svg){: width="400" height="65" loading="lazy"}
+
```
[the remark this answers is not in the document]
MichaIng: The problem with these kind of screenshots is that they either need to be large (larger than possible within the HTML container between navigation and TOC) to be readable. Else one can barely see any content. I'll see if I can make one with small browser window which is sufficiently readable. Thanks for the hind about the hardware page, I just added it there as well with download link and logo.

docs/install.md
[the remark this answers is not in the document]
StephanStS: There is no need to be able to check every click via an image.
But having "optical synchronization points" from time to time is a worthful guidance.

docs/install.md
MichaIng: As said above, the ID is only shown on one of 7 tabs during VM creation. Not sure whether it is reasonable to show only the first of these tabs to refer to it, or showing more than 7 (for the other steps) screenshots, hardly readable or only showing a crop of the browser window?

docs/install.md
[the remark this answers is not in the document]
StephanStS: These screen are not meant to contain readable information. The idea is to have a nice eyecatcher which shows the professional UI and which makes you eager to use the software. :-)

docs/install.md
MichaIng: Same as above, I'll see if I can create a sufficiently small screenshot or crop to keep it readable.

docs/install.md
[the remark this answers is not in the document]
MichaIng: Understood. I personally find those sometimes more distracting, especially when the actual steps one needs to read are then with large space between each other, requiring scrolling and prone to overlook. Also small/bad resolution images to me are often a counter advertisement 😄. But as said, I think a small crop with high DPI (so everything is readable and clear) would be an option.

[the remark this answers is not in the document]
MichaIng: Screenshot added

docs/install.md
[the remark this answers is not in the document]
MichaIng: 1 screenshot added

docs/install.md
MichaIng: Screenshot added

MichaIng:
Btw, I checked again, without success, for an appliance export method. I also examined the backups: This `.vma` format sadly is not a regular tarball but an own archive format with entirely own specifications. As long as we do not want to install Proxmox utilities on the CI/CD runner, we cannot create those. The content however is extremely simple: a raw image and a small config file. I then read that in the past, Proxmox actually did use regular tarballs, not sure why they chose to switch to an own format. Tried to package the image + config into a gzipped tarball. It shows up below backups but cannot be restored. Now trying with an uncompressed tarball. Takes ages as I'm running Proxmox on a USB stick and fiddling with the LVM volume sizes. The tarball has the full raw disk image size, i.e. 8 GiB. But the VM was started already, probably it is smaller or sparse when using the original QCOW2, or in general a compressed QCOW2 instead of an uncompressed raw image.

docs/install.md
```diff
@@ -335,35 +335,35 @@ Select the following tabs for the installation description of your target.
...
-    - Access the Proxmox web interface via HTTPS on TCP port **8006**:
+    1. Access the Proxmox web interface via HTTPS on TCP port **8006**:
         - URL: `https://<your.IP>:8006`
         - Username: `root`
         - Password: `<root user password you entered during Proxmox VE install>`
-    - Select the **Create VM** button at the top right corner. Choose a **Node** , a **VM ID** and a **Name**, then click **Next**.
+    1. Tab **General**: Select the **Create VM** button at the top right corner. Choose a **Node** , a **VM ID** and a **Name**, then click **Next**.
 
-        ![Proxmox VM creation](assets/images/proxmox2.png){: width="722" height="314" loading="lazy"}
+        ![Proxmox VM creation](assets/images/proxmox2.png){: width="600" height="260" loading="lazy"}
```
MichaIng, line 344: Note that the image width is adjusted automatically to fit into the container. Not sure what the benefit is to reduce it manually in this case? Scaling always implies worse quality, and especially on large screens we should make use of the full image size, keeping it readable and not loosing quality.

docs/install.md
```diff
@@ -335,35 +335,35 @@ Select the following tabs for the installation description of your target.
...
-    - While DietPi runs with less, depending on the software you want to install and run within the VM, we recommend at least 1024 MiB memory size, 2048 MiB allows the DietPi system to setup itself without a swap file by default. When done, click **Next**.
+    1. Tab **OS**: Select "**Do not use any media** ", for the **Guest OS** assure that **Linux** and version "**5.x - 2.6 Kernel**" is selected, then click **Next**.
+    1. Tab **System**: As **Machine** you can select `q35`, but the older default will work as well. We recommend the para-virtualised **VirtIO SCSI** controller, which should be the default. Click **Next**.
+    1. Tab **Disks**: Delete the default `scsi0` disk with the red trash bin button, then click **Next**.
+    1. Tab **CPU**: Adjust CPU details as required, we recommend to use the default `kvm64` type. Then click **Next**.
+    1. Tab **Memory**: While DietPi runs with less, depending on the software you want to install and run within the VM, we recommend at least 1024 MiB memory size, 2048 MiB allows the DietPi system to setup itself without a swap file by default. When done, click **Next**.
```
MichaIng, line 352: Good idea to prefix it with the tab name 👍.

Numbering is incremented automatically? I hope it doesn't look confusing when the parent `<font size="+2">` numbering and this list numbering. I'll rebuild the test page so we can look at it.

MichaIng: Looks fine, numbering works fine 👍.

docs/install.md
```diff
@@ -372,16 +372,25 @@ Select the following tabs for the installation description of your target.
...
+        ```sh
         sha256sum -c < <(mawk '/SHA256/{print $2"  DietPi_Proxmox-x86_64-Bullseye.qcow2"}' hash.txt)
-        # Import the disk image
+        ```
+
+        As next, the disk image is imported.  
+        **Note**: Use the VM ID entered during the VM creation.
+
+        ```sh
         ID=100 # replace "100" with the VM ID entered on VM creation
```
MichaIng, line 387:
```suggestion
        **Note**: Replace `100` below with the **VM ID** entered on VM creation.

        ```sh
        ID=100
```

StephanStS (approved):
Added some beautification via PR.
Added comments regarding screenshot size.

docs/install.md
MichaIng: But this highly depends on screen/browser window width. In your particular case it may look nicer, but on larger screens it looks much worse. So generally I'd vote for using the native size to keep the quality and readability of the image at least on larger screens. Of course we could use a smaller crop of the image, but this frame cannot be made smaller, so we'd need to crop off parts left and/or right, which to me doesn't look nice either.

MichaIng: I checked back the style on different screen sizes. Actually the maximum content width (between nav and TOC) is 883 pixels on screens > 1920 (full HD), with . On 1920px screens it's 767px, on smaller screens, first the nav, then the TOC is collapsed, allowing a content width of max 962px at exactly 1219px browser window width.

It is quite easy to apply a max-width for images of 95%, so they are not wider than text (considering line breaks with full words). So to have non-resized images on most screen, we shouldn't use images larger than 725px native size, where possible. Even on >full HD screens the image cannot be that much larger.

I'll apply the 95% max-width rule, so we see whether the layout looks fine with it, without manually capping the image dimensions.

PR #813, opened 2022-12-29
Image size optimization

StephanStS:


StephanStS:
This PR shall be finished after PR #811 is done.

MichaIng:
Great, that makes sense! Let's see if ImgBot can optimise it a little more.

PR #817, opened 2022-12-29
DietPi-Sync and DietPi-Backup descriptions extended

StephanStS:


docs/dietpi_tools.md
```diff
@@ -762,6 +774,16 @@ dietpi-backup
...
 
+=== "Scripted run"
+
+    DietPi-Backup can be run from the command line or from scripts without user interaction by calling it via 
+    
+    ```sh
+    dietpi-backup 1
+    ```
+
+    A similar restore procedure is not recommended to avoid accidentally system overwrites.
```
MichaIng, line 785: Indeed not recommended to do this in non-interactive script, cron jobs or such, but shall we add the info for completeness, e.g. just to skip the clicks through the menu and call the restore directly from console?
```suggestion
    A similar restore procedure is not recommended to avoid accidentally system overwrites. But to skip navigating through the menu, it can be done as well via

    ```sh
    dietpi-backup -1
    ```
```

docs/dietpi_tools.md
```diff
@@ -779,30 +805,98 @@ dietpi-backup
...
+
+    DietPi-Sync can be run from the command line or from scripts without user interaction by calling it via 
+    
+    ```sh
+    dietpi-sync 1
+    ```
+
+=== "Dry run"
+
+    Each sync includes a leading dry run, after which you can check the expected result before deciding if you want to continue with the actual sync:
```
MichaIng, line 893: Unrelated to this PR, but shall we actually keep the forced dry run + space check here? We removed it in `dietpi-backup` since it can take quite long and the space check may show imprecise free space results on network drives, and a failing sync because of insufficient free space is usually not critical. Instead we added an option to enable/disable the dry run, which we could do here as well.

MichaIng (approved):

MichaIng:
Great complementation!

docs/dietpi_tools.md
StephanStS: I also thought about it to keep it in sync with dietpi-backup. But as long as it is not implemented, the docu shall describe the actual status.
If this will be changed in dietpi-sync, the docu will also be changed in parallel. :-)

StephanStS:
> Great completion of those tools' docs!

Maybe we should think about a better structuring of this page. It seems to be quite large and therefore confusing resp. difficult to navigate.

MichaIng:
Or remove the partly primary TOC navigation and have only the tools themselves as primary navigation, sorted e.g. like in `dietpi-launcher`?

docs/dietpi_tools.md
MichaIng: I'll implement the change/alignment with next release.

StephanStS:
Or insert a further main tab "DietPi Tools" to navigate similar to "Optimized Software" (resp. "Software options"):
grafik
and then start the major topics on the left: 
- [DietPi launcher](http://localhost:8000/docs/dietpi_tools/#dietpi-launcher)
- [Software installation](http://localhost:8000/docs/dietpi_tools/#software-installation)
- [System configuration](http://localhost:8000/docs/dietpi_tools/#system-configuration)
- [System maintenance](http://localhost:8000/docs/dietpi_tools/#system-maintenance)
- [Misc tools](http://localhost:8000/docs/dietpi_tools/#misc-tools)

and add the remaining navigation on the right side.

docs/dietpi_tools.md
StephanStS: Then pls don't forget an issue for the Docs. :-)

MichaIng:
Also a good idea, but means that a lot of internal links need to be edited + shared links are invalid 🤔.

StephanStS:
> Also a good idea, but means that a lot of internal links need to be edited + shared links are invalid thinking.

Might be, but nothing is for free. :-)
I propose to do the invest.

PR #821, opened 2023-01-03
Update UrBackup docs

MichaIng:


docs/software/cloud.md
```diff
@@ -351,60 +351,92 @@ Clients are available for Windows, macOS, Linux and FreeBSD.
...
+        ```sh
+        systemctl restart urbackupsrv
+        ```
 
-    - Select `Settings`.
-    - Change the Backup Storage Path (`/mnt/dietpi_userdata/urbackup` is the default path on DietPi).
-    - Click `Save`.
-    - Restart service with `systemctl restart urbackupsrv`.
+    **Best practice: Disable "Allow client-side pausing of backups"**  
+    Backup procedures sometimes are paused by the client side which leads to a long backup duration. To avoid this the pausing can be deactivated via the UrBackup web interface:
```
MichaIng, line 372: Where is the info from that this is best practice? Usually I'd expect the defaults to be best practice. If this feature exists, probably there can be good reasons for clients to pause the process? I mean the feature won't exist for clients to pause or just for fun. And also backup duration is usually not critical, no reason to force a hurry, I'd say. Cannot really find any further info about it in docs.

StephanStS: I did many tests with UrBackup in the last days: 
- UrBackup servers: NanoPi R6S, NanoPi NEO2, Rock Pi S (i.e. from slow, 500 MB RAM to fast, 4 GB RAM)
- UrBackup clients: Win7 VM, Win10 VM, Win11 VM, Win10 on i7 PC (also from quite slow to fast machines)

In all these combinations (i.e. also on NanoPi R6S server with Win10 on i7 client) the system seemed to "hang", so that the backup duration would have last several weeks (!). With the option deselected, all worked within < 1 h.  
From these results I named this a kind of best practice. :-)

StephanStS: Attached one example screenshot 😆 
grafik

I assume I will not live to see it finishing. 
Also, the money for the energy costs are not affordable: Approx 10 Bill$ (@20W, 0,2$/kWh) 🤣

MichaIng: Sounds like something which should be reported to the UrBackup devs or asked in UrBackup forum, i.e. a bug or misbehaviour, or it might be a setup issue which could be even solved without denying clients to pause.

I think we can disable the setting directly in `dietpi-software`. Either there is a `/var/urbackup` file to have it set on first service start or otherwise we can set it in the database directly.

However, good to have the hint here until it is solved in another way.

PR #858, opened 2023-03-06
WordPress: Hint for initial dialog added

StephanStS:
See DietPi Forum entry: [Wordpress setup fails](https://dietpi.com/forum/t/wordpress-setup-fails/16140).


docs/software/social.md
```diff
@@ -135,6 +135,10 @@ It allows you to create your own website.
     - Database Host = `localhost`
     - Table Prefix = `wp_`
 
+    **Remark:** You need to fill out every dialog field although it seems like they are already set (in the example screenshot the global application password is `dietpi`):
+
+    ![WordPress database initialization dialog](../assets/images/dietpi-software-social-wordpress-initialdialog.jpg){: width="350" height="274" loading="lazy"}
```
MichaIng, line 140: Let's use these nice boxes, if we already have them 🙂:
```suggestion
    !!! info "Actively fill all fields"
        You need to fill out every dialog field although it seems like they are already set. In the example screenshot the global application password is `dietpi`.

        ![WordPress database initialization dialog](../assets/images/dietpi-software-social-wordpress-initialdialog.jpg){: width="350" height="274" loading="lazy"}
```

docs/software/social.md
```diff
@@ -135,6 +135,13 @@ It allows you to create your own website.
     - Database Host = `localhost`
     - Table Prefix = `wp_`
 
+    !!! note "Actively fill all fields!"
```
MichaIng, line 138: Isn't `note` adding this pencil icon? I think the exclamation mark of `info` fits better here.

StephanStS: `info` gives an 'i', no '!'.
See https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types

MichaIng: Ah right, however it is more an eye catcher, fitting better for info about solving potential problems. IMO.

StephanStS: @MichaIng: In case of an eyecatcher, I propose this:

image

MichaIng: This is the other extreme, going into "danger" direction. IMO the "info" is the right thing in the middle.

StephanStS: I throwed a coin and you won for this time. :-)

MichaIng: Maybe we should declare some standard regarding these boxes:
- "note" for info which can be interesting or an extension, but doesn't affect whether something succeeds or not
- "tip" for similar info, but such we want to further highlight, because they are pretty cool or contributed by others
- "info" for cases where one/something may not proceed/succeed when not taking into account
- "warning" for cases where one might damage or loose something when not taking into account
- There is also the "danger" box in red, which we do not currently use, IMO it isn't more highlighting than the warning, and not sure how to best differentiate between "warning" and "danger" boxes then.

PR #974, opened 2024-02-07
Add YDNS link to dietpi-ddns section of software installation page.

TDuffinNTU:
Just a small one-liner since I forgot to add this to the docs back when I added support a while back. Thanks!


MichaIng:
Ah right, that one was missing, many thanks!

PR #1038, opened 2024-12-07
Proxmox installation docs updated to ISO file installation

StephanStS:
Issue #1036


MichaIng:
I would actually promote this script as the 1st/recommended method, also I do not see the QCOW2 image import as deprecated at all, as it is faster, less overhead overall. What about this:
1. Describe QCOW2 method via script
  - Add the manual VM creation + image import as by default closed info box.
2. Describe ISO method as 2nd/alternative option

Or do you think, people prefer the ISO method, as it does not require any console access, even that it means ~doubled download size, the manual VM creation steps, and additional time to install DietPi from the ISO to the virtual disk? I mean the script is just a one line command on the Proxmox host console, then all guided whiptail menus, and no additional installation or any manual VM creation. But if you think people still prefer the ISO way, then at least I would not call the QCOW2 method deprecated, in invert script and manual VM creation steps, so that the main method is calling the script only, and the lengthy manual VM generation is in an info box that is closed by default.

LOGIN-TB:
for me the Installation with the VM-ISO files is the goto method. You don't need to install aditional packages on the proxmox server and you are free to configure all settings during the process. And its the recommended way

MichaIng:
The only package that needs to be installed (and is automatically by the script), is `xz-utils`. And the ISO way is recommended by whom/where? I mean just in general: Why would one 1. manually create a VM, 2. download a larger ISO 3. go through a ISO to virtual disk installation process, and alternatively one can run a single command on the console, that asks you for the most important VM parameters which a whiptail menu, with sensible defaults, which can be changed afterwards any time?

I guess it turns down to the question, whether someone who uses Proxmox is just used to create VMs manually, and import ISOs, so that our first mentioned method better aligns to how all other distros provide their Proxmox images.

LOGIN-TB:
Feel free to read the official Proxmox documentation, which explains how to create a virtual machine:

https://pve.proxmox.com/pve-docs/pve-admin-guide.html#chapter_virtual_machines

Although there’s a reason Proxmox provides a GUI for managing the system and its VMs, you’re welcome to suggest any approach you prefer. Keep in mind, however, that while there are many different ways to achieve the same goal, each additional method or script can increase the potential for errors and the need for additional support.

LOGIN-TB:
roxmox’s documentation regarding VM creation is incomplete. There is no step-by-step guide. However, anyone who has ever created a VM via the “Create VM” menu knows that it’s impossible to create a new VM without selecting an ISO! Through the GUI, that is the only way to create a new VM. I assume that anyone running Proxmox knows how to create a new VM. It’s only necessary to point out important, specific settings, such as enabling the “Qemu Agent” so that the tools are installed.

Since Proxmox also provides an API, most (though not all) commands for creating a VM can be executed via the CLI or API. Because this is not self-explanatory, it is described in detail in the documentation. The “user path,” however, is through the GUI.

OVA templates exist because Proxmox wants to attract VMware customers, given Broadcom’s new company and product policies. This option appeared relatively soon after the announcement of the new products and prices, as did the import of ESXi disks.

From my own experience, I can say that handling OVA templates is very cumbersome and not an alternative for VM creation. It’s more like a convoluted workaround rather than a genuine solution—definitely not an alternative.

LOGIN-TB:
I need to correct myself. Of course, it is possible to create a VM without an ISO. However, the default setting is to select an ISO, and an ISO is then required for the OS installation.

StephanStS:
Why using the ISO way? It is the way, Proxmox users are typically used to when they install other VMs.
They might not be Debian specialists who want to deal with the command line interface, because this needs some more knowledge about Linux, about the Proxmox files location (where are the VM storage files located), etc.

With the ISO and the Proxmox GUI the VM generation might need some more steps compared to the script, but the complexity for the typical Proxmox user via ISO and GUI seems to be nuch lower for me.
As an example, my ISO storage looks like this (shall give an estimation that the ISO option is quite often used):
image

What could be done is not to define the script and qcow option as deprecated, instead we should name it "alternative way".

MichaIng:
> I need to correct myself. Of course, it is possible to create a VM without an ISO. However, the default setting is to select an ISO, and an ISO is then required for the OS installation.

Okay that is probably reason enough.

> What could be done is not to define the script and qcow option as deprecated, instead we should name it "alternative way".

Yeah, as it is a perfectly functional and supported way (I mean CLI usage and using virtual disk images), probably as alternative, also outside of the info box with a dedicated header, like "Alternative: Import virtual disk image via console" or so, and in addition to the existing text, the one liner:
```sh
bash <(curl -sSfL https://raw.githubusercontent.com/dazeb/proxmox-dietpi-installer/main/dietpi-install.sh)
```
And then have/keep the long manual description in the info box. Wait, I'll add a code suggestion, as it is only a few lines.

obn100:
Hej Team, if I may - I vote for the ISO. Please have a look at the other operating systems (Free&OpenBSD/Debian/Ubuntu...NixOS). For a long time, everybody basically provides ISO files that were burned on CDs, later used on USB sticks and now are conveniently just added via the dialogue in Proxmox. So that ISO still could be used to be burned on a CD(DVD?) or written to a USB stick. I don't see a drawback in this conformity.

User perspective:
- To be frank: if one OS (derivate) does not provide me with an ISO, I just try a different one
- why should I install a tool for a one-time install? Especially xz which now has a bad reputation despite it was fixed in March 2024 https://en.wikipedia.org/wiki/XZ_Utils_backdoor

docs/install.md
```diff
@@ -452,7 +464,89 @@ Select the following tabs for the installation description of your target.
...
-        Note: Keep the Proxmox VM option "QEMU Guest Agent" option inactive (uncheck check box) to be able to control the VM via the Proxmox GUI.
+        Note: Keep the Proxmox VM option "QEMU Guest Agent" option active (check check box) during the first run of the VM so that the QEMU Guest Agent package is installed within the VM. This enables the control of the VM via the Proxmox GUI.
+
+    ??? info "Alternative installation method by importing the disk image (QCOW2 image)"
+        Following is an alternative installation description where a `.xz` based disk image of format `.qcow2` was used. It contains command line based procedure steps instead of the Proxmox GUI supported VM creation with an `.iso` file described above.
+
+        !!! info "Proxmox Helper Script to install DietPi in Proxmox"
+            There is a script from `Darren Bennett` ([GitHub `dazeb`](https://github.com/dazeb)) which makes the installation steps described below much easier by execution of the installation steps within the script.  
+            See <https://github.com/dazeb/proxmox-dietpi-installer> for details.
+
```
MichaIng, line 475:
````suggestion

<h2>Alternative: Import virtual disk image via console</h2>

An alternative installation method is importing a virtual disk image (QCOW2 format) directly into a newly generated VM. This skips the additional installation process when using the ISO image, but requires some manual commands on the Proxmox OS console. To simplify the process, there is a script from `Darren Bennett` ([GitHub `dazeb`](https://github.com/dazeb)) which allows to enter the essential parameters via terminal UI and creates the VM and volume for you. See <https://github.com/dazeb/proxmox-dietpi-installer> for details.

The script can be executed with this command from the Proxmox OS console:
```sh
bash <(curl -sSfL https://raw.githubusercontent.com/dazeb/proxmox-dietpi-installer/main/dietpi-install.sh)
```

??? info "Click here if you want to do the VM creation and image import manually"
````

MichaIng:
> despite it was fixed in March 2024

and never made it into any stable Debian version ;). But I know what you mean, reputation is damaged in any case.

MichaIng:
Btw, companion PR for the website is this, which adds the ISO image downloads to all VMs: https://github.com/MichaIng/DietPi-Website/pull/345

docs/install.md
MichaIng: I'll merge this, to see how it looks. Generally, as people read through the upper headers anyway before reaching here (and the TOC shows the headers of the RPi installation only, many parts could be skipped or shortened. At some point, we need to split the installation instructions for different hardware/platforms into own pages. Using these tabs does not work well for larger content like used here, breaking TOC links and header numbers. But in a dedicated PR, of course.

MichaIng:
Now it looks like this, better separated from the 3rd header, otherwise it looks like it belongs to "3. Execute the Clonezilla based DietPi installation":
image

Since the header has no number, it does not break the numbering.

@StephanStS one thing I recognised:
image

"Download the `.iso`..." is in a list, but there is no 2. in this section. Instead more nested non-numeric lists start, which are hence quite much indented. I suggest to remove the parent list completely, and optionally give the below dot list numbers. Alternatively:
1. Connect to Proxmox web UI ...
2. Insert the URL ...
3. Press the button ...
4. etc

So to have all steps of that section in one numbered list, like it was/is in the QCOW2 description.

PR #1111, opened 2025-09-24
Add popup/tooltip messages to graphics

StephanStS:
Added tooltips to graphics to enhance the docs UX.

Rule: No tooltips for logos.


MichaIng (approved):
Many thanks, also the old forum screenshot replacement was highly overdue.

PR #1120, opened 2025-10-16
Releasenotes v9.18

StephanStS:


StephanStS:
@MichaIng : Should we move the point "DietPi-LetsEncrypt -> The OCSP toggle has been removed,..." to a new headline "Remarks" or "Notable hints" because it is not really a bugfix?

PR #1127, opened 2025-11-07
Releasenotes v9.19 and new package docu for BirdNET-Go

StephanStS:
@MichaIng : 

- Open point in the RNs is the graphic
- Could you please check the BirdNET-Go docu parts? I tried to deduct it from dietpi-software and the BirdNET-Go web page

I also already updated <https://github.com/MichaIng/DietPi/wiki/DietPi-Software-list> and <https://github.com/MichaIng/DietPi/wiki/DietPi-TCP-UDP-port-usage-list>: Please check the software list entry (entry `#127`).

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -592,6 +592,8 @@ RSA
 RSS
 rsync
 Rsyslog
+RTC
+RTCs
 rTorrent
 runtime
 ruTorrent


StephanStS:
@MichaIng : 
- If you add a new graphic, we can finish the RNs. The contents seems to be o.k.
- Do you think, we should also add the "View logs" tab to Prowlarr, Jackett, Transmission, Deluge, qBittorrent

Issue #1130, opened 2025-11-20
Rework ADS-B documentation to also cover ship tracking and weather balloon tracking


StephanStS:
Rework this docu section: https://dietpi.com/docs/software/distributed_projects/#ads-b-feeder
- add AIS based ship tracking (162 MHz)
- add sondehub weather ballon tracking (400 .. 406 MHz)

PR #1132, opened 2025-11-23
ADS-B docu enhancement: Add info about ship/balloon tracking

StephanStS:
Issue #1130


MichaIng (approved):
Nice, I like the subtitle/description change, so people who do not know these technical term abbreviations know that it is about.

PR #1133, opened 2025-11-24
Remove obsolete upgrade instructions

JappeHallunken:
The docs still contain info about how to upgrade from Debian Stretch to Buster.

This PR will remove the instructions and replace it with links to the corresponding blog posts for the currently by DietPi supported Debian versions.

### DIFF
--- a/docs/assets/stylesheets/custom.css
+++ b/docs/assets/stylesheets/custom.css
@@ -112,12 +112,12 @@
   display: table;
   table-layout: auto;
 }
-#logging-output-filtering-options ~ div div.md-typeset__table table {
-  table-layout: fixed;
-}
-
-#logging-options ~ div div.md-typeset__table table {
-  table-layout: auto;
+table tbody tr td:first-of-type code {
+  /* table-layout: fixed; */
+  white-space: pre-wrap;
+  word-break: keep-all;
+  overflow-wrap: normal;
+  hyphens: none;
 }
 
 /* Footer */


MichaIng:
As said, leaving this in the docs is redundant and bloats this page unnecessarily: That time, also now, the usage/howto page was for dumping misc things we did not set up or find a proper place for yet. Everything there should at best be moved to a matching place:
- Logging actually belongs to https://dietpi.com/docs/software/log_system/, though long term makes more sense in `dietpi-config` or a `dietpi-syslog` script, especially after someone found time to replace `rsyslog` with persistent `systemd-journald` logging => `/var/log/journal`.
- Automation/`dietpi.txt` instructions are needed in (and linked from) install pages. Maybe we split out the long and tested install page at some point and find a way to integrate automation instructions around there. The snippets we started to use give us more possibilities as well.

Hence I am for removing the distro upgrade section there. If we add access to those scripts to `dietpi-update` at some point, we can link the blog posts in the `dietpi-update` docs section, but in usage/howto it looks stray, as the whole page 😅.

MichaIng:
> move the logging stuff to [https://dietpi.com/docs/dietpi_tools/software_installation/#quick-selections-log-system](https://dietpi.com/docs/dietpi_tools/software_installation/#quick-selections-log-system), IMO it doesn't make sense in the software section. Maybe even remove the logging choice from the `dietpi-software` and include into `dietpi-config`?

Agreed. `rsyslog` is basically obsolete, and `logrotate` not worth a software ID in `dietpi-software`. My aim on logging options:
- `dietpi-ramlog #1` default like now
- `dietpi-ramlog #2` with hourly concatenation like now
- Persistent `/var/log` but non-persistent system logs: `journalctl`
- Persistent `/var/log` and persistent system logs: this is automatically the case if `/var/log/journal` exists, which is then used by `systemd-journald`. But if we have a toggle already, makes sense to set `Storage=persistent` explicitly in `/etc/systemd/journald.conf` or better `/etc/systemd/journald.conf.d/dietpi.conf`. and then `Storage=volatile` for the other options explicitly.

For persistent `/var/log` it makes sense to have `logrotate` installed. Since Trixie, wtmp/lastlog log files do not exist anymore, unless their replacements are explicitly installed, but APT and DPkg are still logging there. But otherwise it is only about toggling `dietpi-ramlog.service` and changing configs, hence does not make much sense in `dietpi-software`.

StephanStS:
@JappeHallunken : Could you review my reformatting? I changed the heading levels and indentation to have it fit to other pages.
From my point of view we could merge this PR to the dev branch ("Merge" and not "Squash and merge"). :+1: 

I also updated/upgraded the blog posts, especially [Why you should upgrade your Stretch system now](https://dietpi.com/blog/?p=1001) by using the text you moved out from the docs.

BTW: I also updated the distro tables in 
- [DietPi – How to upgrade to “Bullseye”](https://dietpi.com/blog/?p=811)
- [Debian Bookworm has been released](https://dietpi.com/blog/?p=3128)
- [Debian 13 Trixie has been released](https://dietpi.com/blog/?p=4014)

PR #1204, opened 2026-06-08
Docs - Correct the script path for Home Assistant Update

paulgeisler:


MichaIng:
If the HA dir is at `/home/homeassistant`, then you did not reinstall HA since DietPi v10.2. It was moved to `/opt/homeassistant`, indeed.

Issue #1224, opened 2026-09-10
Update Proxmox VM installation description via .qcow2


StephanStS:
Prerequisite: https://dietpi.com/downloads/images/testing/DietPi_VM-UEFI-ARMv8-Trixie.qcow2 is moved to images

Text passage to be updated: Go to 
https://dietpi.com/docs/install/#how-to-install-dietpi-proxmox
Search for expandable box "Click here if you want to do the VM creation and image import manually"

**--> old:**
apt update
apt full-upgrade
apt install xz-utils
curl -O https://dietpi.com/downloads/images/DietPi_Proxmox-x86_64-Bookworm.qcow2.xz
sha256sum -c <(curl -sSf 'https://dietpi.com/downloads/images/DietPi_Proxmox-x86_64-Bookworm.qcow2.xz.sha256')
xz -d DietPi_Proxmox-x86_64-Bookworm.qcow2.xz

**--> updated:**
apt update
apt full-upgrade
curl -O https://dietpi.com/downloads/images/DietPi_Proxmox-x86_64-Bookworm.qcow2
sha256sum -c <(curl -sSf 'https://dietpi.com/downloads/images/DietPi_Proxmox-x86_64-Bookworm.qcow2.sha256')

PR #1231, opened 2026-09-13
Update install.md with qcow2 and newer Debian version (Bookworm -> Trixie)

StephanStS:
See also #1224


docs/install.md
```diff
@@ -148,17 +148,17 @@ Select the following tabs for the installation description of your target.
              ```sh
              mkdir ~/Downloads
              cd ~/Downloads
-             curl -LO 'https://download.virtualbox.org/virtualbox/7.0.10/virtualbox-7.0_7.0.10-158379~Debian~bookworm_amd64.deb'
-             apt install ./virtualbox-7.0_7.0.10-158379~Debian~bookworm_amd64.deb
+             curl -LO 'https://download.virtualbox.org/virtualbox/7.2.16/virtualbox-7.2_7.2.16-174877~Debian~trixie_amd64.deb'
+             apt install ./virtualbox-7.2_7.2.16-174877~Debian~trixie_amd64.deb
```
MichaIng, line 152: We should switch to use the VirtualBox APT repo. See "Debian-based Linux distributions" here: https://www.virtualbox.org/wiki/Linux_Downloads

Some steps there are outdated, like the `gpg --dearmor`. The `.asc` file can just be used as is. And it must NOT be stored to `/usr/share/keyrings`, which is reserved for packages, not for manual writes. `/etc/apt/keyrings` is the place for custom keys.

But not necessarily needed in this PR.

Also interesting: https://packages.debian.org/sid/virtualbox
But it was never merged into any production suite, just kept in Sid/unstable.

MichaIng (approved):

PR #1233, opened 2026-09-16
fix MariaDB root auth docs (#1222)

### LINKED ISSUE #1222: Fix MariaDB quick access docs
The MariaDB `root` account has no password, but uses the UNIX user socket authentication plugin (by default), which allows the UNIX `root` user to login without password, but no other user can login with this account.

It is intended only to manage per-service/software databases and users, not to be used directly by services/software:
- https://dietpi.com/forum/t/25468
- https://dietpi.com/docs/software/databases/#mariadb-quick-access

JappeHallunken:
Closes #1222


docs/software/databases.md
```diff
@@ -30,10 +30,19 @@ Source: [MariaDB](https://mariadb.com/), [LGPL](https://commons.wikimedia.org/w/
 
 === "Quick access"
 
-    As `root` user, run `mariadb` from command line, no separate authentication required. But note that this won’t work via sudo, but an interactive root user shell session is required.
+    As the Linux `root` user, run `mariadb` from the command line. By default, the MariaDB `root` account uses `unix_socket` authentication, so no MariaDB password is required.
```
MichaIng:
```suggestion
    Run `sudo mariadb`, or `mariadb` as UNIX `root` user from the command line. By default, the MariaDB root user account has no password assigned. Instead, authentication is coupled with the UNIX root user.
```

docs/software/databases.md
```diff
@@ -30,10 +30,19 @@ Source: [MariaDB](https://mariadb.com/), [LGPL](https://commons.wikimedia.org/w/
...
 
-    - Username = `root`
-    - Password = The same as your root login password, default is `dietpi`
+    * Username = `root`
+    * Password = None
+    * Authentication = `unix_socket`
```
MichaIng:
```suggestion
```

docs/software/databases.md
```diff
@@ -30,10 +30,19 @@ Source: [MariaDB](https://mariadb.com/), [LGPL](https://commons.wikimedia.org/w/
...
+
+    The MariaDB `root` account is intended for administrative tasks such as managing databases and database users. Services and software should use dedicated database users instead.
+
+    > **Note:** Run `mariadb` from an interactive Linux `root` shell rather than directly via `sudo`:
+    ```
+    sudo -i
+    mariadb
+    ```
```
MichaIng:
```suggestion
```

MichaIng (changes requested):
I checked and tested again how things are now (on Trixie and Forky):
- While the UNIX user coupling remains true, the `mysql.user` table `plugin` field actually contains `mysql_native_password` now, regardless whether an actual password is assigned or not, and whether the UNIX user coupling is active. In earlier versions it did show `unix_socket` instead. Hence, while the general information about how authentication works is true, we should not mention `unix_socket` as term, which might cause confusion.
- `sudo mariadb` and `sudo mysql` works well now with other users. This did not work when I tested it some years/Debian versions ago, hence looks like the mechanism changed under the hood, too. The note about that hence can be removed.

MichaIng (approved):

